PR #167, opened 2023-01-10
Change smallest_power_of_ten to -64 for floats.

deadalnix:
`18446744073709551615e-65` rounds to `0`, and it is the largest possible value with exponent `-65`. Therefore, we can assume the minimal exponent for which we don't round to zero actually is `-64`.

### DIFF
--- a/include/fast_float/float_common.h
+++ b/include/fast_float/float_common.h
@@ -530,7 +530,7 @@ inline constexpr int binary_format<double>::smallest_power_of_ten() {
 }
 template <>
 inline constexpr int binary_format<float>::smallest_power_of_ten() {
-  return -65;
+  return -64;
 }
 
 template <> inline constexpr size_t binary_format<double>::max_digits() {


lemire:
Thanks !!!

deadalnix:
Merge conflict resolved.

deadalnix:
I rebased on top of the right branch, sorry for the confusion. Is it good to merge? if not, what's missing?

lemire:
@deadalnix  I don't think that there was necessarily a problem with the PR... it did not seem necessary to sync. I do need to carefully check the changes to make sure that we don't introduce bugs. I expect to eventually merge this PR.

tgross35:
@lemire where do these values come from? I am trying to figure out a way to determine them programatically. 

`largest_power_of_ten` seems to be `floor(exponent_bias / log2(10))`, which returns 38 for f32 and 308 for f64. This makes sense (can't represent an exponent larger than the max binary exponent) and does line up with the literals in this repo.

`smallest_power_of_ten` makes less sense. I _think_ that `-(exponent_bias + mantissa_bits)/log2(10)` (possibly +/- an off by one error somewhere). Which seems to make sense because the smallest positive exponent should be the minimum exponent plus the additional places won by representing a subnormal, converted from base2 to base10. However, this gives values of -45 and -323; way off from -65 and -342 present in this repository.

My calculations do seem to align with http://weitz.de/ieee/, e.g. 1e-45 is representable as a f32 (0x1), but 1e-46 gets rounded to 0.

lemire:
@tgross35 

> where do these values come from?

Sure. Please see the papers where this is discussed in details.

- [Fast Number Parsing Without Fallback](https://arxiv.org/abs/2212.06644), Software: Practice and Experience 53 (7), 2023
- [Number Parsing at a Gigabyte per Second](https://arxiv.org/abs/2101.11408), Software: Practice and Experience 51 (8), 2021

[The Python script is part of this repository](https://github.com/fastfloat/fast_float/blob/main/script/table_generation.py).

tgross35:
The paper does say:

> We have that w × 10<sup>−343</sup> < 2<sup>−1074</sup> for all w < 2^<sup>64</sup>.
Thus if the decimal exponent is smaller than -342, then the number must be considered to be zero. If the decimal
exponent is greater than 308, the result must be infinite (beyond the range).

But it doesn't mention the significance of these numbers, at least that I can find. 1074 is bias + mantissa bits and 64 seems to be the same constant for both f32 and f64. My guess is that is to account for possible shift within the internal u64 exponent representation.

Going off of that:

- q<sub>min,shifted</sub> = -⌊(exponent_bias + mantissa_bits + internal_rep_bits) / log2(10)⌋
- q<sub>max,shifted</sub> = ⌊(exponent_bias + 1) / log2(10)⌋

Which gives the following:

- -26..4 for f16. Largest (4) matches https://github.com/fastfloat/fast_float/issues/148, but the smallest value there (-27) is one less than the calculated smallest (-26)
- -59..38 for f16. Largest (38) matches https://github.com/fastfloat/fast_float/issues/148, but the smallest value there (-60) is one less than the calculated smallest (-59)
- -64..38 for f32. Largest (38) matches what is in this repo. Smallest (-64) matches what is in this PR
- -342..307 for f64. Largest (308) and smallest (-342) both match what is in this repo.

I have not tested against the f16 or bf16 algorithms. Does what is proposed here seem accurate?

(cc @jakubjelinek from #148, how did you come up with the values there? And is it possible that -26 and -59 could be the correct min powers of 10?)

lemire:
> > We have that w × 10−343 < 2−1074 for all w < 2^64. Thus if the decimal exponent is smaller than -342, then the number must be considered to be zero. If the decimal
exponent is greater than 308, the result must be infinite (beyond the range).
>
> But it doesn't mention the significance of these numbers, at least that I can find. 

I am not sure what you mean by "doesn't mention the significance of these numbers".

Let me be explicit:
- `10**309` is beyond the range of binary64 and is thus infinite. From the paper: 
Screenshot 2024-06-27 at 5 02 51 PM

- `2**-1074` is effectively zero  in binary64. From the paper:
Screenshot 2024-06-27 at 5 01 30 PM



> Does what is proposed here seem accurate?



Can you please explicitly relate your comments to @deadalnix 's PR.

tgross35:
I do see the numbers in the paper, but I do not see a derevation or relationship with `{largest,smallest}_power_of_ten`, or a way to calculate what those numbers are based on properties. The bit you cite says that the smallest positive value for binary64 is 2<sup>-1074</sup> = 2<sup>bias + mantissa_bits - 1</sup>, that part makes sense.

The problem: ceil(log10(2^-1074)) is ~_-324_, which should be the "smallest power of 10". However, `smallest_power_of_ten` in this repository, the python file, and the paper, is the magic number _-342_. Which I have not seen explained.

I am trying to explain mathematically why this is specifically -342, not -324 or -341 or -343 or anything else. "10e-341" rounds to 0, as does everything smaller until ~"10e-324".

So I am working backwards to try to understand this. As I quoted above from "Fast Number Parsing":

> We have that w × 10<sup>−343</sup> < 2<sup>−1074</sup> for all w < 2^<sup>64</sup>.
Thus if the decimal exponent is smaller than -342, then the number must be considered to be zero. If the decimal
exponent is greater than 308, the result must be infinite (beyond the range).

This allows us to arrive at `-⌊(exponent_bias + mantissa_bits + magic_number_64) / log2(10)⌋`. Because `magic_number_64` needs to be the same for both f32 and f64 (i.e. it is not bit width, plugging in "32" provides an inaccurate answer for `f32`), I am suggesting that `magic_number_64` may be the width of the internal exponent repr (a `uint64_t` in https://github.com/fastfloat/fast_float/blob/1fc3ac3932220b5effaca7203bb1bb771528d256/include/fast_float/decimal_to_binary.h#L94)

---

The relationship to this PR is that it should be trivial to verify the suggested change if such a derevation of the magic number exists. My above formula can produce the number in this PR, saying the change should be correct; I am asking if my formula is correct.

The relation to everything else is aiding in applying this algorithm new float types that are not specified in the paper. E.g. if my above formula is indeed correct, then GCC may have an incorrect (but still functional) `smallest_power_of_ten` for f16 and bf16.

lemire:
@tgross35 If you believe that there is a bug in GCC, please report the issue with GCC. At this time, `fast_float` does not yet support 16-bit floats. This pull request by @deadalnix is not related to 16-bit floats.

Your comments are misplaced. Please take them where they belong.

Let me be clear. This PR is not about 16-bit floats and we are not currently supporting 16-bit floats.

GCC *does* support 16-bit floats, but that is the wrong place to discuss issues with GCC. Please raise the issue where it belongs.

lemire:
@deadalnix Merged. Thanks.

PR #176, opened 2023-02-13
Proof of concept: constexpr from_chars

leni536:
Hi,

On the weekend I played around with trying to make from_chars constexpr, at least when compiled in C++20.

Would such efforts be welcome?

This looks surprisingly straightforward in C++20. The major obstacles are:
1. Work around the rounding mode detection, which can't work in constexpr. This can be managed with `std::is_constant_evaluated()`, either by unconditionally going to the slow path in constexpr, or relying that compilers implement rounding to nearest in constexpr. I implemented the latter, but in hindsight I should probably have implemented the former, just to be safe.
2. `memcpy`. This can be replaced with `std::bit_cast` in C++20.

I have ideas to work around both of these before C++20, but that's a little bit more involved.

### DIFF
--- a/include/fast_float/ascii_number.h
+++ b/include/fast_float/ascii_number.h
@@ -12,9 +12,11 @@ namespace fast_float {
 
 // Next function can be micro-optimized, but compilers are entirely
 // able to optimize it well.
-fastfloat_really_inline bool is_integer(char c)  noexcept  { return c >= '0' && c <= '9'; }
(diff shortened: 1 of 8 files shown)


leni536:
Thanks. I will close this PR and I would approach this incrementally, in multiple PRs:
1. First with low hanging fruit functions that can be made constexpr with minimal change at least from C++14.
2. Then with the more complex functions that need C++20 for std::is_constant_evaluated().
3. Then looking for opportunities to remove the C++20 requirement.

PR #198, opened 2023-04-26
Add support for other char types, opt-in SIMD, JSON parse rules

mayawarrier:
- Add support for other char types (with opt-in SSE2 support to accelerate parsing)
- Add JSON parse rules support


lemire:
@mayawarrier  Can you sync with the main branch?

Also I recommend breaking down the PR. If you want to extend for author syntaxes like JSON, please do so as a separate PR from SIMD optimizations... Otherwise it gets too difficult to review.

Make the PR as small as you can.

mayawarrier:
@lemire I've synced with the main branch and reduced the PR to just the SIMD optimizations, as requested. Awaiting your review!

include/fast_float/float_common.h
```diff
@@ -78,6 +78,28 @@
 #endif
 #endif
 
+#ifdef FASTFLOAT_SSE2
```
lemire, line 81: Above, we would like something like this...

```C++
#if !defined(FASTFLOAT_SSE2) && (defined(__x86_64)  || defined(__x86_64__) || defined(_M_X64) || defined(__amd64) || defined(_M_AMD64))
#define FASTFLOAT_SSE2
#endif
```

Or something of the sort so that SSE2 support is picked up automagically.

lemire:
I am concerned that currently, in this PR, the SSE2 code does not get compiled at all...

mayawarrier:
> I am concerned that currently, in this PR, the SSE2 code does not get compiled at all...

I've added automatic detection of SSE2, so hopefully it should be compiled and tested now.

lemire:
@mayawarrier

Good. Let us run tests. If/when it passes, I will need to do some hand inspection because it has to do with performance and I must make sure we are not messing anything.

lemire:
On my todo to review. I will eventually merge this.

mayawarrier:
@lemire  I made a small perf change. Can you run the tests again?

lemire:
Running tests.

Note that I still lack benchmarking for this, which I will need to add.

I won't add an optimization without some quantification of the benefits.

mayawarrier:
@lemire I attempted to benchmark this using a modified version of the simple_fastfloat_benchmark repo, this was one of the runs on Linux (WSL) using gcc 11.3:

benchmark_results_cropped
I'm seeing on average 10-15% increase in speed for the char16_t SIMD version.

With MSVC however the SIMD version shows a decrease in speed:

benchmark_results2_cropped

include/fast_float/ascii_number.h
```diff
@@ -241,29 +323,31 @@ parsed_number_string_t<UC> parse_number_string(UC const *p, UC const * pend, par
...
         exponent = end_of_integer_part - p + exp_number;
-      } else { // We have a value with a fractional component.
-          p = answer.fraction.ptr;
-          UC const * frac_end = p + answer.fraction.len();
-          while((i < minimal_nineteen_digit_integer) && (p != frac_end)) {
-            i = i * 10 + uint64_t(*p - UC('0'));
-            ++p;
-          }
-          exponent = answer.fraction.ptr - p + exp_number;
+      }
```
mayawarrier, line 342: Sorry about these formatting changes! They were made automatically by my IDE

lemire:
Ok. I have effectively merged your code changes into https://github.com/lemire/simple_fastfloat_benchmark (I did not quite do a PR because I wanted to preserve the project)....

I went to an Ice Lake server (GCC11). I did this...

```
cmake -B build
cmake --build build
./build/benchmarks/benchmark -f data/canada.txt 
```

And I got...
```
# read 111126 lines 
ASCII volume = 1.93374 MB 
netlib                                  :   251.28 MB/s (+/- 29.4 %)    14.44 Mfloat/s      30.55 i/B   557.46 i/f (+/- 0.0 %)     12.01 c/B   219.14 c/f (+/- 1.8 %)      2.54 i/c      3.16 GHz 
doubleconversion                        :   224.78 MB/s (+/- 0.5 %)    12.92 Mfloat/s      54.10 i/B   987.16 i/f (+/- 0.0 %)     13.52 c/B   246.61 c/f (+/- 0.6 %)      4.00 i/c      3.19 GHz 
strtod                                  :   160.91 MB/s (+/- 0.7 %)     9.25 Mfloat/s      70.85 i/B  1292.81 i/f (+/- 0.0 %)     18.88 c/B   344.55 c/f (+/- 0.7 %)      3.75 i/c      3.19 GHz 
abseil                                  :   409.27 MB/s (+/- 2.3 %)    23.52 Mfloat/s      29.36 i/B   535.72 i/f (+/- 0.0 %)      7.43 c/B   135.53 c/f (+/- 2.2 %)      3.95 i/c      3.19 GHz 
fastfloat                               :   871.76 MB/s (+/- 2.3 %)    50.10 Mfloat/s      16.55 i/B   302.00 i/f (+/- 0.0 %)      3.49 c/B    63.62 c/f (+/- 2.3 %)      4.75 i/c      3.19 GHz 
UTF-16 volume = 3.86749 MB 
doubleconversion                        :   461.12 MB/s (+/- 1.9 %)    13.25 Mfloat/s      27.11 i/B   989.16 i/f (+/- 0.0 %)      6.59 c/B   240.47 c/f (+/- 1.9 %)      4.11 i/c      3.19 GHz 
fastfloat                               :  1382.45 MB/s (+/- 1.2 %)    39.72 Mfloat/s      10.34 i/B   377.51 i/f (+/- 0.0 %)      2.20 c/B    80.22 c/f (+/- 1.2 %)      4.71 i/c      3.19 GHz 
```

I did changed the `CMakeLists.txt` file as follows...

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f72f6a6..980fc4b 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,7 @@ endif()
 
 
 FetchContent_Declare(fast_float
-     GIT_REPOSITORY https://github.com/fastfloat/fast_float.git
+     GIT_REPOSITORY https://github.com/mayawarrier/fast_float
      GIT_TAG origin/main
      GIT_SHALLOW TRUE)
```

And I did...

```
cmake -B buildnew
cmake --build buildnew
./buildnew/benchmarks/benchmark -f data/canada.txt 
```

and I got...

```
# read 111126 lines 
ASCII volume = 1.93374 MB 
netlib                                  :   251.94 MB/s (+/- 27.6 %)    14.48 Mfloat/s      30.55 i/B   557.46 i/f (+/- 0.0 %)     12.05 c/B   219.95 c/f (+/- 1.4 %)      2.53 i/c      3.18 GHz 
doubleconversion                        :   223.22 MB/s (+/- 0.6 %)    12.83 Mfloat/s      54.10 i/B   987.16 i/f (+/- 0.0 %)     13.62 c/B   248.58 c/f (+/- 0.5 %)      3.97 i/c      3.19 GHz 
strtod                                  :   161.15 MB/s (+/- 11.3 %)     9.26 Mfloat/s      70.85 i/B  1292.81 i/f (+/- 0.0 %)     18.87 c/B   344.37 c/f (+/- 11.2 %)      3.75 i/c      3.19 GHz 
abseil                                  :   404.10 MB/s (+/- 1.2 %)    23.22 Mfloat/s      29.36 i/B   535.72 i/f (+/- 0.0 %)      7.52 c/B   137.16 c/f (+/- 1.2 %)      3.91 i/c      3.19 GHz 
fastfloat                               :   870.16 MB/s (+/- 1.3 %)    50.01 Mfloat/s      16.51 i/B   301.21 i/f (+/- 0.0 %)      3.49 c/B    63.67 c/f (+/- 1.3 %)      4.73 i/c      3.18 GHz 
UTF-16 volume = 3.86749 MB 
doubleconversion                        :   430.57 MB/s (+/- 0.6 %)    12.37 Mfloat/s      27.11 i/B   989.16 i/f (+/- 0.0 %)      7.06 c/B   257.71 c/f (+/- 0.5 %)      3.84 i/c      3.19 GHz 
fastfloat                               :  1333.69 MB/s (+/- 0.6 %)    38.32 Mfloat/s      10.38 i/B   378.62 i/f (+/- 0.0 %)      2.28 c/B    83.16 c/f (+/- 0.6 %)      4.55 i/c      3.19 GHz 
```

It is not very conclusive.

The `canada.txt` file should be a good test for this because it has many long sequences.

Admittedly, maybe I screwed up somewhere...

mayawarrier:
@lemire I ran the test again (using your version of simple_fastfloat_benchmark, just in case) but I still see the 10-15% increase in comparison to the original on WSL. I even tried varying the order the tests ran in (maybe caching?) but the results were the same.

Then I thought maybe WSL has something to do with it? But a Cygwin build showed the same thing:
```
$ benchmarks/./benchmark -f ../data/canada.txt
# read 111126 lines
volume = 2.03972 MB
char_type = char
netlib                                  :   272.72 MB/s (+/- 4.8 %)    14.86 Mfloat/s      67.30 ns/f
doubleconversion                        :   206.44 MB/s (+/- 4.9 %)    11.25 Mfloat/s      88.91 ns/f
strtod                                  :    63.67 MB/s (+/- 5.8 %)     3.47 Mfloat/s     288.29 ns/f
fastfloat                               :   675.18 MB/s (+/- 7.4 %)    36.78 Mfloat/s      27.19 ns/f
fastfloat_simd                          :   648.62 MB/s (+/- 11.1 %)    35.34 Mfloat/s      28.30 ns/f


volume = 4.07944 MB
char_type = char16_t
doubleconversion                        :   391.28 MB/s (+/- 3.1 %)    10.66 Mfloat/s      93.82 ns/f
fastfloat                               :  1064.43 MB/s (+/- 3.1 %)    29.00 Mfloat/s      34.49 ns/f
fastfloat_simd                          :  1205.15 MB/s (+/- 4.5 %)    32.83 Mfloat/s      30.46 ns/f
```

The processor I'm running this on is an i7-8750H (Coffee Lake). I looked at the latency, throughput, and uops for the SIMD instructions I used (https://www.agner.org/optimize/instruction_tables.pdf), but they appear identical between Ice Lake and Coffee Lake. One thing that is different though is the estimated clock speed from the benchmark, my processor hits almost 4GHz during the test (this shows on CPUz and Task Manager as well), see below:

```
# read 111126 lines
ASCII volume = 2.03972 MB
netlib                                  :   272.27 MB/s (+/- 2.9 %)    14.83 Mfloat/s      29.59 i/B   569.44 i/f (+/- 0.0 %)     13.79 c/B   265.35 c/f (+/- 0.9 %)      2.15 i/c      3.94 GHz
doubleconversion                        :   210.77 MB/s (+/- 3.4 %)    11.48 Mfloat/s      55.18 i/B  1061.94 i/f (+/- 0.0 %)     17.80 c/B   342.63 c/f (+/- 1.2 %)      3.10 i/c      3.93 GHz
strtod                                  :   144.96 MB/s (+/- 6.6 %)     7.90 Mfloat/s      66.12 i/B  1272.61 i/f (+/- 0.0 %)     25.81 c/B   496.69 c/f (+/- 3.2 %)      2.56 i/c      3.92 GHz
abseil                                  :   349.16 MB/s (+/- 4.3 %)    19.02 Mfloat/s      29.65 i/B   570.73 i/f (+/- 0.0 %)     10.74 c/B   206.79 c/f (+/- 1.7 %)      2.76 i/c      3.93 GHz
fastfloat                               :   741.07 MB/s (+/- 11.5 %)    40.37 Mfloat/s      16.75 i/B   322.45 i/f (+/- 0.0 %)      5.00 c/B    96.14 c/f (+/- 9.5 %)      3.35 i/c      3.88 GHz
UTF-16 volume = 4.07944 MB
doubleconversion                        :   406.03 MB/s (+/- 4.3 %)    11.06 Mfloat/s      27.61 i/B  1062.95 i/f (+/- 0.0 %)      9.24 c/B   355.82 c/f (+/- 2.4 %)      2.99 i/c      3.94 GHz
fastfloat                               :  1318.67 MB/s (+/- 10.7 %)    35.92 Mfloat/s      10.23 i/B   393.63 i/f (+/- 0.0 %)      2.84 c/B   109.27 c/f (+/- 8.4 %)      3.60 i/c      3.93 GHz
```

My guess is that the difference in our results is down to some variation in our processors. It could also be that the assembly generated by our compilers are different (I'm using GCC 11.3.0 Ubuntu 11.3.0-1ubuntu1~22.04).

Is it possible to benchmark on different processors to see if there's any difference in the result? Maybe the optimization can be activated for certain processors or environments.

lemire:
In one instance, you are reporting performance counters (instructions per byte and so forth) and in the other instance you are not. WSL won't support performance counters (at least I never managed to get it to do it with me), which is problematic here because the reason SIMD ought to improve the performance has to do with reduced instruction count.

I do not see a reduction in the number of instructions. So something is off.

Here is the same thing on an older processor (AMD Zen2) with GCC11 (Ubuntu 22).

Main:

```
$ ./build/benchmarks/benchmark -f data/canada.txt
# read 111126 lines 
ASCII volume = 1.93374 MB 
netlib                                  :   250.62 MB/s (+/- 0.7 %)    14.40 Mfloat/s      31.21 i/B   569.44 i/f (+/- 0.0 %)     12.89 c/B   235.20 c/f (+/- 0.6 %)      2.42 i/c      3.39 GHz 
doubleconversion                        :   182.96 MB/s (+/- 1.7 %)    10.51 Mfloat/s      56.01 i/B  1021.94 i/f (+/- 0.0 %)     17.64 c/B   321.94 c/f (+/- 1.7 %)      3.17 i/c      3.38 GHz 
strtod                                  :   136.55 MB/s (+/- 1.6 %)     7.85 Mfloat/s      69.74 i/B  1272.61 i/f (+/- 0.0 %)     23.65 c/B   431.45 c/f (+/- 1.6 %)      2.95 i/c      3.39 GHz 
abseil                                  :   375.14 MB/s (+/- 0.3 %)    21.56 Mfloat/s      30.35 i/B   553.72 i/f (+/- 0.0 %)      8.61 c/B   157.13 c/f (+/- 0.2 %)      3.52 i/c      3.39 GHz 
fastfloat                               :   785.45 MB/s (+/- 0.2 %)    45.14 Mfloat/s      16.93 i/B   309.00 i/f (+/- 0.0 %)      4.11 c/B    75.06 c/f (+/- 0.1 %)      4.12 i/c      3.39 GHz 
UTF-16 volume = 3.86749 MB 
doubleconversion                        :   362.47 MB/s (+/- 0.8 %)    10.41 Mfloat/s      28.03 i/B  1022.95 i/f (+/- 0.0 %)      8.91 c/B   325.13 c/f (+/- 0.7 %)      3.15 i/c      3.39 GHz 
fastfloat                               :  1242.90 MB/s (+/- 0.2 %)    35.71 Mfloat/s      10.54 i/B   384.51 i/f (+/- 0.0 %)      2.60 c/B    94.83 c/f (+/- 0.1 %)      4.05 i/c      3.39 GHz 
```

This PR:

```
$ ./buildnew/benchmarks/benchmark -f data/canada.txt
# read 111126 lines 
ASCII volume = 1.93374 MB 
netlib                                  :   259.52 MB/s (+/- 0.5 %)    14.91 Mfloat/s      31.21 i/B   569.44 i/f (+/- 0.0 %)     12.44 c/B   227.06 c/f (+/- 0.4 %)      2.51 i/c      3.39 GHz 
doubleconversion                        :   182.41 MB/s (+/- 0.8 %)    10.48 Mfloat/s      56.01 i/B  1021.94 i/f (+/- 0.0 %)     17.70 c/B   322.95 c/f (+/- 0.8 %)      3.16 i/c      3.39 GHz 
strtod                                  :   138.46 MB/s (+/- 1.0 %)     7.96 Mfloat/s      69.74 i/B  1272.61 i/f (+/- 0.0 %)     23.32 c/B   425.52 c/f (+/- 1.0 %)      2.99 i/c      3.39 GHz 
abseil                                  :   347.22 MB/s (+/- 1.5 %)    19.95 Mfloat/s      30.35 i/B   553.72 i/f (+/- 0.0 %)      9.30 c/B   169.75 c/f (+/- 1.3 %)      3.26 i/c      3.39 GHz 
fastfloat                               :   792.75 MB/s (+/- 0.2 %)    45.56 Mfloat/s      16.89 i/B   308.21 i/f (+/- 0.0 %)      4.07 c/B    74.35 c/f (+/- 0.1 %)      4.15 i/c      3.39 GHz 
UTF-16 volume = 3.86749 MB 
doubleconversion                        :   382.90 MB/s (+/- 0.3 %)    11.00 Mfloat/s      28.03 i/B  1022.95 i/f (+/- 0.0 %)      8.44 c/B   307.84 c/f (+/- 0.2 %)      3.32 i/c      3.39 GHz 
fastfloat                               :  1201.45 MB/s (+/- 1.0 %)    34.52 Mfloat/s      10.57 i/B   385.62 i/f (+/- 0.0 %)      2.69 c/B    98.02 c/f (+/- 1.0 %)      3.93 i/c      3.38 GHz 
```

Please focus on the `i/B` column. If this column does not show gains (i.e., reductions), then the PR is not likely beneficial.

lemire:
> Or perhaps the user can be given an option to opt-in to SIMD, so they can enable it if it shows an improvement for their workflow.

There are many things that may change... instruction latency, CPU frequency... However, a hard constant is the number of retired instructions. That's typically easy to measure accurately (better than 1% on the same machine, run to run).

You are reporting `10.23` instructions per byte. That's credible. Now is that before or after the PR? If that's the number with the PR, what was the number before? How much gain did you get?


If you are not seeing a significant reduction in the number of instructions retired, then it is probably not a good optimization.

mayawarrier:
@lemire 10.23 is after the PR.

Here's a run on WSL with performance counters showing both versions:
```
# read 111126 lines
volume = 2.03972 MB
char_type = char
netlib                                  :   267.75 MB/s (+/- 4.0 %)    14.59 Mfloat/s      29.59 i/B   569.44 i/f (+/- 0.0 %)     14.04 c/B   270.29 c/f (+/- 2.1 %)      2.11 i/c      3.94 GHz
doubleconversion                        :   217.00 MB/s (+/- 3.4 %)    11.82 Mfloat/s      55.18 i/B  1061.94 i/f (+/- 0.0 %)     17.32 c/B   333.32 c/f (+/- 2.4 %)      3.19 i/c      3.94 GHz
strtod                                  :   144.72 MB/s (+/- 3.8 %)     7.88 Mfloat/s      66.12 i/B  1272.61 i/f (+/- 0.0 %)     26.23 c/B   504.93 c/f (+/- 1.3 %)      2.52 i/c      3.98 GHz
abseil                                  :   410.18 MB/s (+/- 5.8 %)    22.35 Mfloat/s      29.65 i/B   570.73 i/f (+/- 0.0 %)      9.36 c/B   180.13 c/f (+/- 3.0 %)      3.17 i/c      4.03 GHz
fastfloat                               :   792.93 MB/s (+/- 9.1 %)    43.20 Mfloat/s      16.75 i/B   322.46 i/f (+/- 0.0 %)      4.72 c/B    90.82 c/f (+/- 7.1 %)      3.55 i/c      3.92 GHz
fastfloat_simd                          :   696.91 MB/s (+/- 9.9 %)    37.97 Mfloat/s      16.75 i/B   322.45 i/f (+/- 0.0 %)      5.40 c/B   103.89 c/f (+/- 7.5 %)      3.10 i/c      3.94 GHz


volume = 4.07944 MB
char_type = char16_t
doubleconversion                        :   418.18 MB/s (+/- 4.6 %)    11.39 Mfloat/s      27.56 i/B  1060.95 i/f (+/- 0.0 %)      9.12 c/B   351.18 c/f (+/- 2.0 %)      3.02 i/c      4.00 GHz
fastfloat                               :  1077.22 MB/s (+/- 5.3 %)    29.34 Mfloat/s      10.20 i/B   392.51 i/f (+/- 0.0 %)      3.47 c/B   133.68 c/f (+/- 4.4 %)      2.94 i/c      3.92 GHz
fastfloat_simd                          :  1291.45 MB/s (+/- 8.5 %)    35.18 Mfloat/s      10.23 i/B   393.63 i/f (+/- 0.0 %)      2.91 c/B   111.88 c/f (+/- 7.2 %)      3.52 i/c      3.94 GHz
```

 i/B increases, which means it might not be a good optimization, as you said. But how does that explain the increase in speed I see consistently across multiple runs?

include/fast_float/ascii_number.h
```diff
@@ -47,6 +64,40 @@ uint64_t read_u64(const char *chars) {
...
+#ifdef FASTFLOAT_SSE2
+
+fastfloat_really_inline
+uint64_t simd_read8_to_u64(const __m128i data) {
+FASTFLOAT_SIMD_DISABLE_WARNINGS
+  static const char16_t kmasks[] = { 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff };
+  const __m128i masks = _mm_loadu_si128(reinterpret_cast<const __m128i*>(kmasks));
+
+  // todo: with AVX512BW + AVX512VL, can use cvtepi16_epi8 instead of masking + pack
+  __m128i masked = _mm_and_si128(data, masks);
```
lemire, line 76: Why are you masking the data here? If the numbers are all digits, then they are all ASCII, correct? In which case, what is the masking doing?

include/fast_float/ascii_number.h
```diff
@@ -47,6 +64,40 @@ uint64_t read_u64(const char *chars) {
...
+  __m128i packed = _mm_packus_epi16(masked, masked);
+
+  uint64_t val;
```
lemire, line 79: Possibly the store_u could be replace by _mm_cvtsi128_si64x which may not be simpler but I don't think that's even right.

You already have the value in a SIMD register... finish the job right away... something like this...

```
const __m128i ascii0 = _mm_set1_epi8('0');
  const __m128i mul_1_10 =
      _mm_setr_epi8(10, 1, 10, 1, 10, 1, 10, 1, 10, 1, 10, 1, 10, 1, 10, 1);
  const __m128i mul_1_100 = _mm_setr_epi16(100, 1, 100, 1, 100, 1, 100, 1);
  const __m128i mul_1_10000 =
      _mm_setr_epi16(10000, 1, 10000, 1, 10000, 1, 10000, 1);
  const __m128i input = _mm_sub_epi8(
      _mm_loadu_si128(reinterpret_cast<const __m128i *>(chars)), ascii0);
  const __m128i t1 = _mm_maddubs_epi16(input, mul_1_10);
  const __m128i t2 = _mm_madd_epi16(t1, mul_1_100);
  const __m128i t3 = _mm_packus_epi32(t2, t2);
  const __m128i t4 = _mm_madd_epi16(t3, mul_1_10000);
  return _mm_cvtsi128_si32(
      t4); 
```

Well. It is worth investigating.

mayawarrier: I had something similar before in this PR (which I took from http://0x80.pl/articles/simd-parsing-int-sequences.html and modified for SSE2) but I removed it because the improvement wasn't much and I thought it might be difficult to maintain.

Here's a run on WSL2. I see an improvement of 3-4% over multiple runs. 
```
# read 111126 lines
volume = 2.03972 MB
char_type = char
netlib                                  :   270.02 MB/s (+/- 3.0 %)    14.71 Mfloat/s      29.59 i/B   569.44 i/f (+/- 0.0 %)     13.91 c/B   267.69 c/f (+/- 1.5 %)      2.13 i/c      3.94 GHz
doubleconversion                        :   211.86 MB/s (+/- 4.3 %)    11.54 Mfloat/s      55.18 i/B  1061.94 i/f (+/- 0.0 %)     17.70 c/B   340.69 c/f (+/- 3.3 %)      3.12 i/c      3.93 GHz
strtod                                  :   143.99 MB/s (+/- 2.3 %)     7.84 Mfloat/s      66.12 i/B  1272.61 i/f (+/- 0.0 %)     26.02 c/B   500.86 c/f (+/- 1.2 %)      2.54 i/c      3.93 GHz
abseil                                  :   342.49 MB/s (+/- 4.2 %)    18.66 Mfloat/s      29.60 i/B   569.73 i/f (+/- 0.0 %)     10.98 c/B   211.31 c/f (+/- 2.6 %)      2.70 i/c      3.94 GHz
fastfloat                               :   756.29 MB/s (+/- 8.6 %)    41.20 Mfloat/s      16.75 i/B   322.46 i/f (+/- 0.0 %)      4.98 c/B    95.85 c/f (+/- 6.5 %)      3.36 i/c      3.95 GHz
fastfloat_simd                          :   700.84 MB/s (+/- 9.3 %)    38.18 Mfloat/s      16.75 i/B   322.45 i/f (+/- 0.0 %)      5.37 c/B   103.29 c/f (+/- 7.5 %)      3.12 i/c      3.94 GHz
fastfloat_simd2                         :   700.00 MB/s (+/- 8.9 %)    38.14 Mfloat/s      16.75 i/B   322.45 i/f (+/- 0.0 %)      5.37 c/B   103.37 c/f (+/- 6.3 %)      3.12 i/c      3.94 GHz


volume = 4.07944 MB
char_type = char16_t
doubleconversion                        :   414.08 MB/s (+/- 5.4 %)    11.28 Mfloat/s      27.56 i/B  1060.95 i/f (+/- 0.0 %)      9.27 c/B   356.66 c/f (+/- 2.2 %)      2.97 i/c      4.02 GHz
fastfloat                               :  1180.94 MB/s (+/- 9.1 %)    32.17 Mfloat/s      10.20 i/B   392.51 i/f (+/- 0.0 %)      3.24 c/B   124.72 c/f (+/- 6.1 %)      3.15 i/c      4.01 GHz
fastfloat_simd                          :  1290.11 MB/s (+/- 8.2 %)    35.14 Mfloat/s      10.20 i/B   392.65 i/f (+/- 0.0 %)      2.90 c/B   111.82 c/f (+/- 7.1 %)      3.51 i/c      3.93 GHz
fastfloat_simd2                         :  1330.85 MB/s (+/- 6.9 %)    36.25 Mfloat/s      10.23 i/B   393.63 i/f (+/- 0.0 %)      2.82 c/B   108.58 c/f (+/- 6.0 %)      3.63 i/c      3.94 GHz
```

If we include this then we should also add workflows to test it (since it is SSE4.1 and compilers only make SSE2 available by default).

include/fast_float/ascii_number.h
mayawarrier: Oops, that's a mistake. You're right, it's not necessary.

include/fast_float/ascii_number.h
lemire: You are right that we don't expect a large gain, and the fact that it is not basic SSE2 is a serious concern because we have, thus far, stayed away from runtime dispatching in this library. We should only throw more sophistication at the problem if it is warranted by significant gains.

include/fast_float/float_common.h
```diff
@@ -139,6 +170,14 @@ fastfloat_really_inline constexpr bool cpp20_and_in_constexpr() {
 #endif
 }
 
+fastfloat_really_inline constexpr bool has_simd() {
```
lemire, line 173: I think that this function is never used.

include/fast_float/ascii_number.h
```diff
@@ -76,40 +126,80 @@ uint32_t parse_eight_digits_unrolled(uint64_t val) {
...
+  if (cpp20_and_in_constexpr()) {
+    return false;
+  }   
+#ifdef FASTFLOAT_SSE2
+FASTFLOAT_SIMD_DISABLE_WARNINGS
+  const __m128i data = _mm_loadu_si128(reinterpret_cast<const __m128i*>(chars));
+
+  // (x - '0') <= 9
+  // http://0x80.pl/articles/simd-parsing-int-sequences.html
+  const __m128i t0 = _mm_sub_epi16(data, _mm_set1_epi16(80));
```
lemire, line 164:
```suggestion
  const __m128i t0 = _mm_sub_epi16(data, _mm_set1_epi16(-32720));
```

mayawarrier:
@lemire You're right, I redid the math and your numbers should be correct.

mayawarrier:
@lemire You're right! Running this again shows a big difference in i/B:
```
# read 111126 lines
volume = 2.03972 MB
char_type = char
netlib                                  :   269.68 MB/s (+/- 2.7 %)    14.69 Mfloat/s      29.59 i/B   569.44 i/f (+/- 0.0 %)     13.95 c/B   268.47 c/f (+/- 1.3 %)      2.12 i/c      3.94 GHz
doubleconversion                        :   200.68 MB/s (+/- 2.1 %)    10.93 Mfloat/s      55.18 i/B  1061.94 i/f (+/- 0.0 %)     18.69 c/B   359.74 c/f (+/- 1.0 %)      2.95 i/c      3.93 GHz
strtod                                  :   143.64 MB/s (+/- 1.8 %)     7.83 Mfloat/s      66.12 i/B  1272.61 i/f (+/- 0.0 %)     26.07 c/B   501.72 c/f (+/- 0.8 %)      2.54 i/c      3.93 GHz
abseil                                  :   340.77 MB/s (+/- 6.0 %)    18.57 Mfloat/s      29.60 i/B   569.73 i/f (+/- 0.0 %)     11.02 c/B   212.12 c/f (+/- 3.6 %)      2.69 i/c      3.94 GHz
fastfloat                               :   739.06 MB/s (+/- 6.9 %)    40.26 Mfloat/s      16.75 i/B   322.46 i/f (+/- 0.0 %)      5.04 c/B    96.92 c/f (+/- 5.1 %)      3.33 i/c      3.90 GHz
fastfloat_simd                          :   768.20 MB/s (+/- 5.8 %)    41.85 Mfloat/s      16.75 i/B   322.46 i/f (+/- 0.0 %)      4.89 c/B    94.08 c/f (+/- 4.5 %)      3.43 i/c      3.94 GHz
fastfloat_simd2                         :   767.71 MB/s (+/- 9.6 %)    41.83 Mfloat/s      16.75 i/B   322.46 i/f (+/- 0.0 %)      4.90 c/B    94.30 c/f (+/- 5.9 %)      3.42 i/c      3.94 GHz


volume = 4.07944 MB
char_type = char16_t
doubleconversion                        :   398.90 MB/s (+/- 7.1 %)    10.87 Mfloat/s      27.56 i/B  1060.95 i/f (+/- 0.0 %)      9.39 c/B   361.39 c/f (+/- 4.7 %)      2.94 i/c      3.93 GHz
fastfloat                               :  1089.92 MB/s (+/- 17.7 %)    29.69 Mfloat/s      10.20 i/B   392.51 i/f (+/- 0.0 %)      3.46 c/B   133.01 c/f (+/- 15.6 %)      2.95 i/c      3.95 GHz
fastfloat_simd                          :  1408.70 MB/s (+/- 7.2 %)    38.37 Mfloat/s       8.79 i/B   338.34 i/f (+/- 0.0 %)      2.67 c/B   102.90 c/f (+/- 4.9 %)      3.29 i/c      3.95 GHz
fastfloat_simd2                         :  1446.00 MB/s (+/- 7.1 %)    39.39 Mfloat/s       8.48 i/B   326.59 i/f (+/- 0.0 %)      2.59 c/B    99.77 c/f (+/- 4.5 %)      3.27 i/c      3.93 GHz
```

include/fast_float/ascii_number.h
mayawarrier: GCC and clang define the macro \_\_SSE_4_1\_\_ when SSE4.1 is available, so it should just be a matter of checking if it exists and adding the -msse4.1 cmdline option to the workflows (-msse4.1 is required regardless of target). MSVC doesn't provide this macro but we can check for \_\_AVX\_\_ which should imply SSE4.1.  
With the corrected math, I see only a 2-3% improvement though. Should I add it?

mayawarrier:
@lemire Awaiting your review. Is there anything I'm missing?

include/fast_float/ascii_number.h
mayawarrier: I suppose this can be changed to _mm_add_epi16(data, _mm_set1_epi16(32720))

lemire:
Looks like a solid 10% performance boost. I expect to merge this.

Note that sometimes it can take some time because I must carefully review it and run my own manual tests, to make sure we don't cause any problems or regression.

You can assume that this PR will be merged soon.

mayawarrier:
@lemire Should I add my name to the contributors list for this PR?

mayawarrier:
> Looks like a solid 10% performance boost. I expect to merge this.

Just wondering, is it only a 10% boost on your end, or do you mean in terms of i/B? 
The sse2 version seems to show a 29% boost (1404.91 MB/s vs 1089.22 MB/s) from the benchmark I posted above.

lemire:
@mayawarrier I meant when adding the branch in https://github.com/lemire/simple_fastfloat_benchmark

I did not take into account your numbers when I said 10%. I think something might be off with your benchmarking code or methodology. If you reduce the instruction count by 10%, if you are lucky, you can be 10%, but not 30% faster. A 30% speed boost is not credible. 

I am also getting about 10% with NEON:

https://github.com/fastfloat/fast_float/pull/209

lemire:
Note that a 10% speed boost matters.

PR #241, opened 2024-03-08
Use version 6.1.0 instead 6.0.0

carlosal1015:
Hi there, I would like to ask if is possible to update version in order to sent this patch to [downstream](https://gitlab.archlinux.org/archlinux/packaging/packages/fast_float/-/issues) because if this digits are different generates issues, for example for https://aur.archlinux.org/packages/ants#comment-960051 or any project that has `find_package(VTK)` and has `fast_float-6.1.0`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.9)
 
-project(fast_float VERSION 6.0.0 LANGUAGES CXX)
+project(fast_float VERSION 6.1.0 LANGUAGES CXX)
 set(FASTFLOAT_CXX_STANDARD 11 CACHE STRING "the C++ standard to use for fastfloat")
 set(CMAKE_CXX_STANDARD ${FASTFLOAT_CXX_STANDARD})
 option(FASTFLOAT_TEST "Enable tests" OFF)


lemire:
Thanks. I have retagged the release with this fix.

ptrcnull:
hiya, just wanted to let you know that retagging releases causes issues for downstream distribution maintainers ( for example, see https://gitlab.archlinux.org/archlinux/packaging/packages/fast_float/-/issues/1 ) - changing a tag changes the tarball, which then makes all builds fail because of checksum mismatches 😅

in this case, bumping the patch version would have probably been a better choice

lemire:
@carlosal1015 @ptrcnull 

I literally I cannot *update the version* without changing the tar ball and the checksum. This is not possible. Sorry.

There is just no perfect solution. Sorry.

I am not sure why the CMake version wasn't updated. I'll make sure to fix the process in the future.

ptrcnull:
then just.. keep the 6.1.0 with the CMake version mismatch? when you have a bug, you don't fix it by amending the release, you fix it by cutting a new one :p

nekopsykose:
> I literally I cannot update the version without changing the tar ball and the checksum.  

the request wasn't to update the same version and not change the tarball (which is logically impossible indeed), but to just make a new version, like `v6.1.1`. natural numbers are free, after all

lemire:
> the request wasn't to update the same version

Please see above. @carlosal1015  requested that we update the version, it was not to issue a new version. Please see screenshot of the comment above. Please review the PR changes.

Screenshot 2024-03-08 at 6 16 22 PM

I am making sure that this issue does not occur again in the future.

For the time being, this issue is closed.

> natural numbers are free, after all

Please see my comment above: there is no perfect solution in such cases.

PR #303, opened 2025-02-13
Fix -Wundef warnings seen in gcc 4.8

biojppm:


dalle:
I think there are more places in the code base where this is applicable

PR #307, opened 2025-03-05
Improwing speed and reduce code size when fast_float is using as internal parser code.

IRainman:
### 🚀 Performance and Code Size Improvements

* Optimized parsing logic for use as an **internal lightweight parser** (for example, inside other libraries).
* Reduced binary size by conditionally excluding unneeded parsing features.
* The `from_chars_result_t` structure is reduced to **4 bytes** for better memory efficiency.
* Improved the `parsed_number_string_t` layout and increased `constexpr`/`consteval` propagation to enable compile-time optimizations.
* Reduced register and cache pressure and branching in parsing hot paths.
* Improved performance in both 64-bit and 32-bit builds for all supported types.

---

### ⚙️ New Configuration Macros

Introduced new optional macros to minimize overhead when certain parsing features are not required:

* **`FASTFLOAT_ONLY_POSITIVE_C_NUMBER_WO_INF_NAN`**
  Restricts parsing to **positive C-style numbers only** — no sign characters, no `INF`, `INFINITY`, or `NaN` and no additional options like skip white spaces or support for Fortran or JSON.

* **`FASTFLOAT_ONLY_ROUNDS_TO_NEAREST_SUPPORTED`**
  Assumes that only the **IEEE 754 “round-to-nearest”** rounding mode is needed, removing crutches support code for other modes.

* **`FASTFLOAT_ISNOT_CHECKED_BOUNDS`**
  Disables bounds checking for input ranges that are assumed to be valid.
  Use only when inputs are guaranteed safe — this reduces branching and slightly improves performance.

* **`FASTFLOAT_ASSUME`**
  Provides a portable abstraction for the compiler’s `[[assume]]` intrinsic.

* **`FASTFLOAT_HAS_BYTESWAP`**
  Uses std::byteswap if available to reduce code size and speed up.

* **`FASTFLOAT_HAS_BIT_CAST`**
  Uses std::bit_cast if available to reduce code size and speed up.

---

### 🧩 Remove Deprecated Macros

* Removed obsolete macroses **`FASTFLOAT_ALLOWS_LEADING_PLUS`** and **`FASTFLOAT_SKIP_WHITE_SPACE`**, which are now superseded by the stricter limited-mode options above.

---

### 🧠 Internal Refactoring and Quality Improvements

* Simplified internal parsing logic with compile-time branching controlled by the new macros.
* Enhanced `constexpr` coverage across the codebase for greater compile-time evaluation.
* Updated benchmarks and tests to validate new restricted configurations.
* Fixed minor compiler warnings and addressed **PVS-Studio** static analysis feedback.
* Properly use FASTFLOAT_SIMD_DISABLE_WARNINGS and FASTFLOAT_SIMD_RESTORE_WARNINGS only for instructions that allow unaligned loads.
* Many small cleanups and fixes.
* Added more comments about functionality and realization details.

---

### 📘 Documentation and Build

* Updated **README.md** to describe all new configuration macros and their intended use cases.
* Adjusted build scripts and benchmark configurations for the new compile-time flags.

---

✅ **Result:**
Smaller, faster by default, and more configurable builds — now library can be used for **internal numeric parsers** or **embedded environments** — while maintaining full compatibility, functionality, and performance improvements in the default configuration.


dalle:
I recently changed some compile time config options into runtime config options. I believe that both worlds would be good. I would prefer if the compile time options to not be any ifdef-macros, but controlled by using template arguments.

lemire:
In this instance, the argument here is to reduce compiled size and improve runtime performance we avoid checking negatives ('-') and for 'nan'/'inf'. In my view, it is an extraordinary claim that checking for the negative ('-') and for 'nan'/'inf' makes a large difference.

It is possibly, but I'd like to see the numbers. Please note that we have do have benchmarks included in the library,

>  Because current realisation parse options in fully software mode and this is bad.

How did you benchmark to arrive at the conclusion that it is bad?

> I recently changed some compile time config options into runtime config options.

I would not assume that it makes a difference to the performance unless we have a hard data. Some parameters can be passed as template arguments, certainly... 

But we need to take into account that parsing a single number takes hundreds of instructions. Predictable branches are not a big concern.

IRainman:
> I recently changed some compile time config options into runtime config options. I believe that both worlds would be good. I would prefer if the compile time options to not be any ifdef-macros, but controlled by using template arguments.

Very well! I'm currently reworking my patch to use templates argument and want to add more constexpr / consteval to the code.

IRainman:
> In this instance, the argument here is to reduce compiled size and improve runtime performance we avoid checking negatives ('-') and for 'nan'/'inf'. In my view, it is an extraordinary claim that checking for the negative ('-') and for 'nan'/'inf' makes a large difference.
> 
> It is possibly true, but I'd like to see the numbers. Please note that we have do have benchmarks included in the library.
> 

I'm measuring performance on the mathematical parser, that already processes minus signs, and I yesterday rewrote it to also process inf / infinity for mathematical questions. In mathematical parser nan isn't possible at all because if it's nan the parser stops processing equations immediately. 

> > Because current realisation parse options in fully software mode and this is bad.
> 
> How did you benchmark to arrive at the conclusion that it is bad?
> 
Options it's not fully processed by the template and generate many additional lines of code. Also complete removing the minus sign for mantissa and inf/nan from fast_float significantly reduces generated assembler.

To check the performance of such a case, tests should contain only positive numbers in general notation.

dalle:
> > I recently changed some compile time config options into runtime config options. I believe that both worlds would be good. I would prefer if the compile time options to not be any ifdef-macros, but controlled by using template arguments.
> 
> Very well! I'm currently reworking my patch to use templates argument and want to add more constexpr / consteval to the code.

Please don't do this until we have the performance comparisons on the gains.

IRainman:
> > > I recently changed some compile time config options into runtime config options. I believe that both worlds would be good. I would prefer if the compile time options to not be any ifdef-macros, but controlled by using template arguments.
> > 
> > 
> > Very well! I'm currently reworking my patch to use templates argument and want to add more constexpr / consteval to the code.
> 
> Please don't do this until we have the performance comparisons on the gains.

Of course, I just cleaned my code of defines and replaced them with an option. Unfortunately, code isn't fully clean-up at compile time as a preprocessing macro.

IRainman:
P. S. my current version of the test code is: 
`	[[assume(_view._Unchecked_end() - _view._Unchecked_begin() >= 1)]];
	double val;
	constexpr auto options{fast_float::chars_format::general | fast_float::chars_format::no_infnan | fast_float::chars_format::disallow_leading_sign};
	const auto res = fast_float::from_chars(_view._Unchecked_begin(), _view._Unchecked_end(), val, options);
	[[assume(res.ptr - _view._Unchecked_begin() >= 1)]];
	const size_t n = res.ptr - _view._Unchecked_begin();`

include/fast_float/ascii_number.h
```diff
@@ -282,38 +286,39 @@ report_parse_error(UC const *p, parse_error error) {
...
-                    parse_options_t<UC> options) noexcept {
-  chars_format const fmt = detail::adjust_for_feature_macros(options.format);
-  UC const decimal_point = options.decimal_point;
+                    const parse_options_t<UC> options) noexcept {
 
   parsed_number_string_t<UC> answer;
   answer.valid = false;
   answer.too_many_digits = false;
-  // assume p < pend, so dereference without checks;
+  [[assume(p < pend)]]; // assume p < pend, so dereference without checks;
```
dalle, line 294: Assume C++11 compatibility

include/fast_float/float_common.h
```diff
@@ -63,16 +67,24 @@ template <typename UC> struct from_chars_result_t {
...
+  FASTFLOAT_CONSTEXPR20 explicit parse_options_t(chars_format fmt = chars_format::general,
+                                     UC dot = UC('.'), const int b = 10) noexcept
+      : format(fmt), decimal_point(dot), base(static_cast<uint8_t>(b)) {}
 
   /** Which number formats are accepted */
-  chars_format format;
+  const chars_format format
+      // adjust for deprecated feature macros
+#ifdef FASTFLOAT_ALLOWS_LEADING_PLUS
+                            | chars_format::allow_leading_plus
```
dalle, line 78: This won't compile, and even if it did, it will be overridden by `format(fmt)` in the constructor list.

include/fast_float/ascii_number.h
IRainman: Hmm, maybe I don't understand correctly, but it's standard since C++23 https://en.cppreference.com/w/cpp/language/attributes/assume and older versions is produced a warning. I need to place this construction into macros, that depends on the platform?

dalle: Yes, the assume attribute is C++23, but this library supports C++11, so that attribute cannot be used.

include/fast_float/float_common.h
IRainman: Fix for this https://github.com/fastfloat/fast_float/pull/307/commits/f496321570c6ad7e12dd309934217b6eb3c30805

dalle:
I'm sorry to interrupt you, but I think you're going the wrong way here and focusing on the wrong stuff. I don't want to seem pompous or ignorant, and I don't want to offend you, I just don't want you to (possibly) waste more time on small changes.

I will say it clearly: Without proof of the gains there is very low chances that PR will be merged.

So please focus on the tests, make the performance/benchmarks to compile and run. Then you should run the performance/benchmarks workflows. If there are quantified improvements then the likelihood of merging this PR will increase substantially.

lemire:
I agree with @dalle.

It is fine to fork the library for your own purposes… it is open source. You can adapt it to your own needs. Please do so.

But we won’t take this PR unless we are convinced of its benefits. And you have not provided the evidence.

We are concerned about maintaining the code, keeping it correct and well tested. 

Changes must be motivated.

IRainman:
I added an improvement test with parser emulation.

P. S. In real code of mathematical processing engine, my patches also give significant improvements: 

std::from_chars

Tests: exactly: 115, almost: 9,
 failed: 4,
 time is: 32064ms.

fast_float::from_chars

Tests:
 time is: 26684ms.

( commits from this PR )

Tests (100% of CPU cores is used by BOINC):
 time is: 43135ms.
Tests (50% of CPU cores is used by BOINC):
 time is: 28898ms.
Tests (BOINC is stopped):
 time is: 25459ms.

When switch FASTFLOAT_ONLY_POSITIVE_C_NUMBER_WO_INF_NAN
is applied:

Tests: exactly: 115, almost: 9,
 failed: 4,
 time is: 25947ms.

( commits from this PR )

Tests (100% of CPU cores is used by BOINC):
 time is: 39588ms.
Tests (50% of CPU cores is used by BOINC):
 time is: 27301ms.
Tests (BOINC is stopped):
 time is: 24597ms.

IRainman:
Additionally, you can also test the compilation flag FASTFLOAT_ONLY_POSITIVE_C_NUMBER_WO_INF_NAN with a file from here https://population.un.org/wpp/downloads?folder=Standard%20Projections&group=CSV%20format — this is a CSV with only positive numbers in there.

lemire:
@IRainman We already have a dataset made entirely of positive values (mesh), see https://github.com/fastfloat/fast_float/pull/308

If you prepare a file that contains just one number per line, you can then pass it to our benchmark like so...
```
cmake -B build -D FASTFLOAT_BENCHMARKS=ON
cmake --build build
./build/benchmarks/realbenchmark myfile.txt
```

(This assumes Linux/macOS... The instructions are similar under Visual Studio except that you need to specify that you work in release mode.)

IRainman:
Hmm, I tried to update the brunch, and it's successfully updated, but now this PR isn't updated. What do I need to do? Close this and create a new PR after I finish changes, or may be some command for this situation exists?

lemire:
> Hmm, I tried to update the brunch, and it's successfully updated, but now this PR isn't updated. What do I need to do? 

I am not certain of what you mean. The content of a PR is determined by the content of a branch.

IRainman:
> I am not certain of what you mean. The content of a PR a determined by the content of a branch.

Now it's probably fine, thank you. I see my new commit in this PR after.

IRainman:
Well, I don't know how to compile basictest.cpp locally, and currently I'm done because it's not any ability to compile tests locally on Windows. If you can upgrade the tests I will try again.

IRainman:
Now I have added additional improvements, fixed a warning, and made tests in Linux work again.

lemire:
@IRainman We cannot merge this as-is. Please break down your PR into smaller contributions, starting from the less consequential.

include/fast_float/ascii_number.h
```diff
@@ -350,46 +363,50 @@ parse_number_string(UC const *p, UC const *pend,
...
   }
+#endif
   else if (digit_count == 0) { // we must have encountered at least one integer!
     return report_parse_error<UC>(p, parse_error::no_digits_in_mantissa);
   }
-  int64_t exp_number = 0; // explicit exponential part
-  if ((uint64_t(fmt & chars_format::scientific) && (p != pend) &&
-       ((UC('e') == *p) || (UC('E') == *p))) ||
-      (uint64_t(fmt & detail::basic_fortran_fmt) && (p != pend) &&
-       ((UC('+') == *p) || (UC('-') == *p) || (UC('d') == *p) ||
```
IRainman, line 375: @lemire @dalle Can you confirm that code is correct? I didn't thoroughly understand logic here.

IRainman:
image

IRainman:
I don't understand why clang-format is still failing after I formatted all files. What's wrong? 🤷🏻‍♀️ [script for Windows](https://github.com/IRainman/fast_float/blob/main/format.cmd) it's definitely doing that needed.

IRainman:
This job is always failed, but clang-format locally confirmed that is already formatted https://github.com/IRainman/fast_float/actions/runs/14861477644/job/41727207426 I don't understand this it all. This is weird.

IRainman:
I'm probably done, you can review and do some speed tests. Now all tests are working. Can you please give me advice, what test I need to add for my changes?

IRainman:
I make tests on the Linux machine:

[original.txt](https://github.com/user-attachments/files/20741045/original.txt)
[pr307.txt](https://github.com/user-attachments/files/20741047/pr307.txt)
[pr307_with_flags.txt](https://github.com/user-attachments/files/20741046/pr307_with_flags.txt)


image
image

IRainman:
@lemire please refer to the numbers in the comment above this.

IRainman (commented):
Upgraded to 8.1.0

IRainman (commented):
Needs custom check.

IRainman (commented):
Well, today I fixed the last error in the parsing; tomorrow I'll test error code processing in multibyte strings because it's the last problem.

IRainman:
image
@dalle @lemire updated performance comparison.

IRainman:
In the small numbers (ip test) also big improvements 
image

IRainman (commented):
Currently I'm done. All other errors are probably from compilers or from workflows.

IRainman:
Currently I'm done. All other errors are probably from compilers or from workflows. I also fixed some errors that I found when code quality checking. There is a performance comparsion:
image

IRainman:
Progress for today:
image

PR #323, opened 2025-09-10
Fix wording

thbeu:


README.md
```diff
@@ -58,40 +58,44 @@ Example:
...
+    std::cerr << "parsing failure\n";
+    return EXIT_FAILURE;
+  }
+  std::cout << "parsed the number " << result << '\n';
   return EXIT_SUCCESS;
 }
 ```
 
-Though the C++17 standard has you do a comparison with `std::errc()` to check whether the conversion worked, you can avoid it by casting the result to a `bool` like so:
+You can simplify the conversion success check by casting the result of `from_chars` to `bool`, as shown below:
```
alugowski, line 76: The original text sounds clunky but it makes an important point that this edit removes. The C++17 standard does not include this cast to bool. It's important for users to know that it is a fast_float extension.

The cast to `bool` only became standard in C++26. https://en.cppreference.com/w/cpp/utility/from_chars_result.html

thbeu: Yes, that's easier to comprehend for non-native English speakers. So what about C++20 then?

jwakely: Everything else in that document talks about C++17 `from_chars`, and I think it should be obvious that unless stated otherwise, the same API is present in C++20 and C++23.

lemire:
This changes too many lines needlessly. I will revise the readme manually.

README.md
```diff
@@ -58,40 +58,44 @@ Example:
...
+  const std::string input = "3.1416 xyz ";
   double result;
-  if(auto answer = fast_float::from_chars(input.data(), input.data() + input.size(), result)) {
-    std::cout << "parsed the number " << result << std::endl;
+  if (auto answer = fast_float::from_chars(input.data(), input.data() + input.size(), result)) {
+    std::cout << "parsed the number " << result << '\n';
     return EXIT_SUCCESS;
   }
-  std::cerr << "failed to parse " << result << std::endl;
+  std::cerr << "parsing failure\n";
```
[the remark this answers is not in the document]
lemire: @thbeu Point taken Thomas. Thanks.

PR #363, opened 2026-01-28
Clarify behavior for underflow

pitrou:
In https://github.com/fastfloat/fast_float/pull/189 the behavior on underflow was changed to better match the standard's recommendations, but the README does not mention underflow explicitly.


pitrou:
FTR we encountered this in Arrow: https://github.com/apache/arrow/issues/49003

lemire:
Merging.

PR #365, opened 2026-02-02
Fix overzealous build-time  `/permissive-` propagating to compile-time (cross-compilation bug)

### LINKED ISSUE #364: Overzealous `/permissive-` breaks cross-compiled builds
Currently, the usage of `target_compile_options(fast_float INTERFACE /permissive-)` when the host compiler allows `/permissive` (as validated by `CHECK_CXX_COMPILER_FLAG(/permissive- FASTFLOAT_COMPILER_SUPPORTS_PERMISSIVE)`) is overzealous and breaks cross-compiled builds. For example, if the host is built using MSVC, but the downstream uses any non-MSVC compiler, compiling will fail with an error such as this:

```
clang: error: no such file or directory: '/permissive-'
```

This is an extension of #304/#305, which fixed a different edge of this issue.

With the current approach (using `target_compile_options(fast_float INTERFACE /permissive-)`) the compiler flag gets propagated forward into the generated `fast_floats-targets.cmake` at build-time:

```CMake
set_target_properties(FastFloat::fast_float PROPERTIES
  INTERFACE_COMPILE_FEATURES "cxx_std_11"
  INTERFACE_COMPILE_OPTIONS "/permissive-"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
)
```

Which erroneously forces the consumer to use the `/permissive-` flag even if it's not relevant for the downstream compiler.

This issue is particularly relevant when working with fast_float via [vcpkg](https://github.com/microsoft/vcpkg), as vcpkg makes it very easy to set up cross-compiling by building for the host development system with their compiler of choice, and then compile for the target using a different relevant compiler, allowing development for other OS's or to allow special compiler-specific features to be enabled/disabled at build time (in my case, my target builds are done using Clang++ as I need some of its features, but my host uses MSVC for speed and ease of setup).

nyoma-diamond:
Resolves #364

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,13 +57,7 @@ if(FASTFLOAT_SANITIZE)
   endif()
 endif()
 
-include(CheckCXXCompilerFlag)
-unset(FASTFLOAT_COMPILER_SUPPORTS_PERMISSIVE)
-CHECK_CXX_COMPILER_FLAG(/permissive- FASTFLOAT_COMPILER_SUPPORTS_PERMISSIVE)
-
-if(FASTFLOAT_COMPILER_SUPPORTS_PERMISSIVE)
-  target_compile_options(fast_float INTERFACE /permissive-)
-endif()
+target_compile_options(fast_float INTERFACE $<$<AND:$<CXX_COMPILER_ID:MSVC>,$<VERSION_GREATER_EQUAL:$<CXX_COMPILER_VERSION>,19.10>>:/permissive->)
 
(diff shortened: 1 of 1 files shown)


nyoma-diamond:
Ran some experimentation on my side to validate that this works as anticipated for cross-compiling MSVC -> Clang++. Still need to test pre-2017 versions of MSVC (i.e., <1910) that don't support `/permissive-`. Working on validating it later. That said, I don't see any reason why that would fail. I think this should be good to go :)

PR #379, opened 2026-04-16
Replace std::min with ternary operators to avoid <algorithm> dependency

mlippautz:
`fastfloat_strncasecmp` relies on `std::min`.

### DIFF
--- a/include/fast_float/ascii_number.h
+++ b/include/fast_float/ascii_number.h
@@ -594,7 +594,7 @@ parse_int_string(UC const *p, UC const *pend, T &value,
           ((digits + 0x46464646u) | (digits - 0x30303030u)) & 0x80808080u;
       uint32_t tz = (uint32_t)countr_zero_32(magic); // 7, 15, 23, 31, or 32
       uint32_t nd = (tz == 32) ? 4 : (tz >> 3);
-      nd = (uint32_t)std::min((size_t)nd, len);
+      nd = (uint32_t)(nd < len ? nd : len);
       if (nd == 0) {
         if (has_leading_zeros) {
           value = 0;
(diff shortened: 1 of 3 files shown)


mlippautz:
Alternatively, maybe it's better to avoid `std::min` in first place to avoid the dependency.

jwakely:
Yes, `<algorithm>` is a large header, especially since all the range overloads got added in C++20.

mlippautz:
Do you prefer just adding a `constexpr size_t fastfloat_min(size_t a, size_t b)` or completely inlining the call?

Context: We'd like to roll to the latest fast_float version in Chrome and ideally avoid a downstream patch.

biojppm:
I agree that the include should be avoided due to its large size.

As for picking between a function or using the ternary operator (which is what I guess you mean by inlining the call?): I guess both are fine, but maybe the latter is preferable, given that std::min is scarcely used. But this is not a strong position.

Grepping for `std::min` in the source, I get
```
File: float_common.h
401:12:      sz = std::min(sz, length - i);

File: ascii_number.h
597:22:      nd = (uint32_t)std::min((size_t)nd, len);

File: digit_comparison.h
112:12:    cb(am, std::min<int32_t>(shift, 64));
```

lemire:
@mlippautz Something like `fast_float::details::min` would be nice.

mlippautz:
> @mlippautz Something like `fast_float::details::min` would be nice.

Sorry, didn't see this comment until I now updated the PR. Shall we go with `fast_float::details::min` or the ternary operators now? :)

lemire:
@mlippautz Ternary is fine.

lemire:
Running tests. I expect them to pass.

lemire (approved):

lemire:
The MINGW32 failure is unrelated.

mlippautz:
@lemire can you help with landing this?

Also, could you cut a release for this as well? We set up auto rollers now for this repository. As long as things are compatible we will semi automatically get the latest changes.

lemire:
Yes, of course.

PR #383, opened 2026-06-01
Parallelize the exhaustive float32 sweeps across hardware threads (~75-88x)

fcostaoliveira:
`exhaustive32_midpoint` sweeps all 2³² float bit-patterns one at a time, which takes
~30 min on a fast machine. The values are independent, so this splits the range across
`std::thread::hardware_concurrency()` threads, with an atomic flag for fail-fast. The
per-value work is factored into a `check_word()` helper; the checks and pass/fail
behavior are unchanged.

On a 96-core machine the runtime drops from ~1900 s to ~25 s (≈75×), identical result:

```
gcc 13  : PASS in 26s
clang 18: PASS in 22s
```

Notes:
- The exhaustive tests are gated behind `FASTFLOAT_EXHAUSTIVE` (off by default and not
  built in CI), so this has no CI impact — it's purely a faster local/dev sweep.
- `tests/CMakeLists.txt` now links `Threads::Threads` (with
  `THREADS_PREFER_PTHREAD_FLAG`).
- C++11, compiles clean under the existing `-Werror -Wall -Wextra -Weffc++
  -Wconversion -Wsign-conversion -Wshadow` set on gcc and clang; clang-format clean.

The same pattern applies to the sibling sweeps (`exhaustive32`, `exhaustive32_64`); I
kept this PR to one test for review, happy to extend it if you'd like.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -4,6 +4,10 @@ cmake_minimum_required(VERSION 3.11 FATAL_ERROR)
 
 include(FetchContent)
 
+# Some tests (the exhaustive sweeps) parallelize across std::thread.
+set(THREADS_PREFER_PTHREAD_FLAG ON)
+find_package(Threads REQUIRED)
+
 option(SYSTEM_DOCTEST "Use system copy of doctest" OFF)
 option(FASTFLOAT_SUPPLEMENTAL_TESTS "Run supplemental tests" ON)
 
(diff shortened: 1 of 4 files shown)


lemire (approved):
This looks good.

fcostaoliveira:
Extended to the sibling sweeps as offered — `exhaustive32` and `exhaustive32_64` now use the same `std::thread` split, preserving each test's existing failure behavior (abort for `exhaustive32`, stop-flag for `exhaustive32_64`).

96-core timings (gcc), all pass and clean under -Werror on gcc + clang:
| test | before | after |
|---|---|---|
| exhaustive32 | ~1055s | 12s |
| exhaustive32_64 | ~1555s | 18s |
| exhaustive32_midpoint | ~1839s | 22s |

Happy to drop the two extra files back out if you'd prefer to keep this PR to just the midpoint test.

lemire:
It is fine. Merging.

PR #386, opened 2026-06-03
Skip materializing parsed_number_string_t spans on the hot path (addresses #384)

fcostaoliveira:
## Summary

`parsed_number_string_t` carries two `span<UC const>` members (`integer`, `fraction`) that are read **only on the rare slow paths** — `digit_comp`, and the `>19`-significant-digit truncation recompute. But they are written on **every** parse, which forces the ~56/64-byte struct to be materialized and marshaled through the by-value return. On the hot path that surfaces as backend/store pressure. This addresses #384 ("the structure `parsed_number_string_t` is ... probably too fat for its own good").

## What changed

- `parse_number_string` gains a **runtime** `bool store_spans = true` parameter (default keeps every existing caller unchanged). When `false`, the `integer`/`fraction` span stores and the span-reading `>19`-digit recompute are skipped.
- `from_chars_float_advanced` parses with `store_spans = false`, attempts Clinger + Eisel-Lemire inline, and routes the two rare slow branches (`too_many_digits`, `am.power2 < 0`) to a single `fastfloat_noinline` (`noinline`+`cold`) helper that re-parses **with** spans and calls the unchanged `from_chars_advanced`.
- New `fastfloat_noinline` macro in `float_common.h`.

Two deliberate choices:
- **Runtime flag, not a template parameter** — a template would create a *second* instantiation of the whole scanner whose icache cost wipes out the gain.
- **`noinline cold` slow path** — the rare re-parse must stay out of line, or the force-inlined hot scanner gets duplicated into the caller; that bloats the hot frame and lengthens the loop-carried dependency chain, which regresses some targets (notably ARM gcc) even though it removes the spill.

Public `from_chars` / `from_chars_advanced` / `parsed_number_string_t` are unchanged.

## Performance

Per-parser microbench (`from_chars<double>` in a tight loop over each dataset, median of 5, pinned core). **base → this PR, MB/s (Δ%)**, vs current tip:

| Target | random | canada | mesh |
|---|---|---|---|
| ARM Neoverse-V2 (Graviton4) gcc | 1087 → 1907 (+75%) | 948 → 1645 (+73%) | 503 → 1489 (+196%) |
| ARM Neoverse-V2 (Graviton4) clang | 1347 → 1449 (+8%) | 1049 → 1135 (+8%) | 879 → 976 (+11%) |
| Intel Ice Lake (Xeon 8360Y) gcc | 1142 → 1358 (+19%) | 973 → 1138 (+17%) | 814 → 955 (+17%) |
[...]

PR #387, opened 2026-06-06
Using unlikely markers for PR386

lemire:
PR https://github.com/fastfloat/fast_float/pull/386 improves performance when parsing short strings. 

However, it does so by marking functions as cold. 

I think that a better option is to make branches as unlikely.

This is what this PR does.

### DIFF
--- a/include/fast_float/float_common.h
+++ b/include/fast_float/float_common.h
@@ -212,18 +212,7 @@ using parse_options = parse_options_t<char>;
 #endif
 
 #ifdef FASTFLOAT_USE_UNLIKELY_ATTR
-// We have to disable -Wc++20-extensions for the [[unlikely]] attribute
-// See comment for @jwakely at
-// https://github.com/fastfloat/fast_float/pull/387#discussion_r3366943539
-// This is unfortunate.
-#ifdef __clang__
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wc++20-extensions"
-#endif
 #define fastfloat_unlikely(x) (x) [[unlikely]]
-#ifdef __clang__
-#pragma clang diagnostic pop
-#endif
 #elif defined(__GNUC__) || defined(__clang__)
 #define fastfloat_unlikely(x) (__builtin_expect(!!(x), 0))
 #else
--- a/include/fast_float/float_common.h
+++ b/include/fast_float/float_common.h
@@ -197,13 +197,25 @@ using parse_options = parse_options_t<char>;
 #define fastfloat_really_inline inline __attribute__((always_inline))
 #endif
 
-// Force a function OUT of line and onto the cold path. Used for the rare
-// slow-path re-parse so the force-inlined hot scanner is not duplicated into
-// the caller (which bloated the hot frame and hurt ILP on some targets).
-#ifdef FASTFLOAT_VISUAL_STUDIO
-#define fastfloat_noinline __declspec(noinline)
+// Branch-probability hint marking the rare slow-path branches as cold, so the
+// optimizer keeps the out-of-line slow-path re-parse off the hot path (and does
+// not duplicate the force-inlined hot scanner into the caller, which bloated
+// the hot frame and hurt ILP on some targets). Used at the call site as
+//   if fastfloat_unlikely(cond) { ... }
+// (the macro supplies the parentheses). It expands to the standard [[unlikely]]
+// attribute in C++20 or newer, otherwise to __builtin_expect on GCC/Clang, or
+// to a no-op elsewhere (e.g. pre-C++20 MSVC, which has no equivalent hint).
+// The [[unlikely]] branch is gated on the language version, not just on
+// __has_cpp_attribute: GCC and Clang report the attribute as available even
+// under -std=c++17, where using it would trip -Wc++20-extensions/-Werror.
+#if (__cplusplus >= 202002L ||                                                 \
+     (defined(_MSVC_LANG) && _MSVC_LANG >= 202002L)) &&                        \
+    defined(__has_cpp_attribute) && __has_cpp_attribute(unlikely) >= 201803L
+#define fastfloat_unlikely(x) (x) [[unlikely]]
+#elif defined(__GNUC__) || defined(__clang__)
+#define fastfloat_unlikely(x) (__builtin_expect(!!(x), 0))
 #else
-#define fastfloat_noinline __attribute__((noinline, cold))
+#define fastfloat_unlikely(x) (x)
 #endif
 
 #ifndef FASTFLOAT_ASSERT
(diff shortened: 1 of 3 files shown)


include/fast_float/float_common.h
```diff
@@ -197,6 +197,27 @@ using parse_options = parse_options_t<char>;
...
+//   if fastfloat_unlikely(cond) { ... }
+// (the macro supplies the parentheses). It expands to the standard [[unlikely]]
+// attribute in C++20 or newer, otherwise to __builtin_expect on GCC/Clang, or
+// to a no-op elsewhere (e.g. pre-C++20 MSVC, which has no equivalent hint).
+// The [[unlikely]] branch is gated on the language version, not just on
+// __has_cpp_attribute: GCC and Clang report the attribute as available even
+// under -std=c++17, where using it would trip -Wc++20-extensions/-Werror.
+#if (__cplusplus >= 202002L ||                                                 \
+     (defined(_MSVC_LANG) && _MSVC_LANG >= 202002L)) &&                        \
+    defined(__has_cpp_attribute) && __has_cpp_attribute(unlikely) >= 201803L
```
jwakely: If `defined(__has_cpp_attribute)` is false then `__has_cpp_attribute(foo)` is ill formed because the undefined macro is replaced with 0 and you get `0(foo)` which is an error.

jwakely: You can avoid that problem with:

```
#if (__cplusplus >= 202002L ||                                                 \
     (defined(_MSVC_LANG) && _MSVC_LANG >= 202002L)) &&                        \
    defined(__has_cpp_attribute)
#if __has_cpp_attribute(unlikely) >= 201803L
#define fastfloat_unlikely(x) (x) [[unlikely]]
#endif
#endif

#ifndef fastfloat_unlikely
#if defined(__GNUC__) || defined(__clang__)
#define fastfloat_unlikely(x) (__builtin_expect(!!(x), 0))
#else
#define fastfloat_unlikely(x) (x)
#endif
#endif
```

include/fast_float/float_common.h
```diff
@@ -197,6 +197,27 @@ using parse_options = parse_options_t<char>;
...
+// Branch-probability hint marking the rare slow-path branches as cold, so the
+// optimizer keeps the out-of-line slow-path re-parse off the hot path (and does
+// not duplicate the force-inlined hot scanner into the caller, which bloated
+// the hot frame and hurt ILP on some targets). Used at the call site as
+//   if fastfloat_unlikely(cond) { ... }
+// (the macro supplies the parentheses). It expands to the standard [[unlikely]]
+// attribute in C++20 or newer, otherwise to __builtin_expect on GCC/Clang, or
+// to a no-op elsewhere (e.g. pre-C++20 MSVC, which has no equivalent hint).
+// The [[unlikely]] branch is gated on the language version, not just on
+// __has_cpp_attribute: GCC and Clang report the attribute as available even
```
jwakely: I know Clang gives that warning, but I don't think GCC does. It's annoying, because it defeats the entire purpose of standardizing things like `__has_cpp_attribute` and feature test macros, which was to get rid of the fragile preprocessor checks for specific versions of specific compilers :sob:

You could put this around the whole of `parse_number.h` where the attribute is used:

```
#ifdef __clang__
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wc++20-extensions"
#endif
...
...
#ifdef __clang__
#pragma clang diagnostic pop
#endif
```

jwakely: That would allow the attribute to be used if `__has_cpp_attribute(unlikely)` is true, without the `__cplusplus >= 202002L` check alongside it.

include/fast_float/float_common.h
jwakely: Alternatively something like this keeps all the messy conditions in one place, separate from the main if-else logic:

```
#ifdef __has_cpp_attribute
#if __has_cpp_attribute(unlikely) /* && any extra conditions you want/need */
#define FASTFLOAT_USE_UNLIKELY_ATTR 1
#endif
#endif

#ifdef FASTFLOAT_USE_UNLIKELY_ATTR
#define fastfloat_unlikely(x) (x) [[unlikely]]
#elif defined(__GNUC__) || defined(__clang__)
#define fastfloat_unlikely(x) (__builtin_expect(!!(x), 0))
#else
#define fastfloat_unlikely(x) (x)
#endif
```

lemire:
@jwakely Much appreciated. I update and ping you again.

lemire:
@jwakely I guess that you have noticed that @fcostaoliveira provided a few significant optimizations. I think GCC should eventually update to benefit from this work.

jwakely:
> @jwakely I guess that you have noticed that @fcostaoliveira provided a few significant optimizations. I think GCC should eventually update to benefit from this work.

Yes, I've already mentioned it to Patrick who did the work to include your code in GCC, thanks :)

include/fast_float/float_common.h
lemire: covered in my last commit

include/fast_float/float_common.h
lemire: covered in my last commit

lemire:
@jwakely I went with your proposals, please have a look.

It is somewhat depressing code, but...

include/fast_float/float_common.h
```diff
@@ -197,6 +197,39 @@ using parse_options = parse_options_t<char>;
...
+#endif
+
+#ifdef FASTFLOAT_USE_UNLIKELY_ATTR
+// We have to disable -Wc++20-extensions for the [[unlikely]] attribute
+// See comment for @jwakely at
+// https://github.com/fastfloat/fast_float/pull/387#discussion_r3366943539
+// This is unfortunate.
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wc++20-extensions"
```
jwakely: I might be wrong, but I think the pragmas have to be placed where the macro is used, not where it's defined :(

(I can't check right now sorry)

jwakely: Oh I see you've already got it at the point of use, so I _think_ you don't need the pragmas here.

fcostaoliveira:
Re-benchmarked the current head (`520fded`) vs `main` across Cascade Lake / Ice Lake / Granite Rapids (gcc 11) and Graviton4 (clang 18), C++17 and C++20: still **+8–24%** on the short-string datasets, and identical to the earlier revision (the C++17 `[[unlikely]]` switch is performance-neutral, within noise). 👍

include/fast_float/float_common.h
lemire: Yeah, I trimmed it out. Thanks.

It is ugly, but what can we do ?

lemire:
@jwakely Eventually (in the future), we shall just require C++20 and all this mess won't be necessary.

PR #399, opened 2026-07-23
Fix compatibility with min()/max() macros

dg0yt:
`windows.h` may define min() and max() macros. It is possible to guard against this on the consuming side. But this might get fragile with transitive dependencies. This PR suggests to avoid the problem in the headers.
Noticed while updating vcpkg ports (breaking build of opencv4 and opencv3).


lemire:
Good.

PR #402, opened 2026-08-04
Avoid unnecessary fractional scans in exact conversion

perfloop-agent:
## Summary

When exact conversion has already found a nonzero discarded integer digit, skip scanning the fractional span for another truncation signal that cannot change rounding. The fractional scan remains in place when the discarded integer suffix is all zero, where it is still needed. The change affects only the rare digit-comparison fallback and adds a focused mode to the existing benchmark for this path.

## Performance

On workload `direct parse_mantissa: 770-digit integer with a nonzero discarded suffix and a 4 MiB zero fraction`, median metric `ns/op` changed from 220801 to 928; paired median delta -219875 (-99.6% of baseline; at least 19/20 confidence interval -226131 to -215201 from 10 pairs).

On workload `public from_chars exact fallback: 770-digit integer with a nonzero discarded suffix and a 4 MiB zero fraction`, median metric `ns/op` changed from 2287640 to 2028719; paired median delta -265772 (-11.6% of baseline; at least 19/20 confidence interval -273683 to -246648 from 10 pairs).

## Testing

Ran the Release CMake test suite, the C++20 constexpr basictest, clang-format 17, and the repository benchmark modes. Focused coverage exercises direct mantissa parsing and public from_chars conversions with zero and nonzero discarded integer suffixes, fractional-tail variants, and the exact-conversion route.

All 4 declared correctness checks passed.

---

Authored and verified by [Perfloop](https://app.perfloop.ai): every claim above was co-measured on both trees and independently re-verified before submission — the full record is public: [case_9k6jbbrrax](https://app.perfloop.ai/t/oss/case_9k6jbbrrax). Replies from this account are human-approved, and a human operator is accountable for this contribution.


include/fast_float/digit_comparison.h
IRainman, line 1: Thank you, I merged this into my version!

PR #404, opened 2026-08-08
docs: fix strtod function name in README

latent-9:
Small typo in the README: it refers to the C string-to-double function as `stdtod`, but the function is `strtod`. (Being a faster `strtod` is the whole point of the library.)


lemire:
Indeed thanks

PR #410, opened 2026-09-21
Keep chars_format::javascript out of the common hot path

Algunenano:
`chars_format::javascript`, new in 8.3.0, adds its ECMAScript leading-zero rule as a runtime test on `fmt` inside `parse_number_string`. Unlike `basic_json_fmt`, which is a template parameter and compiles away, this one cannot: its two error returns and the Annex B octal scan land in the parser body that every `from_chars` instantiation force-inlines, and `fmt` has to stay live across the integer scan.

This turned up while reviewing the v8.3.0 bump in ClickHouse (https://github.com/ClickHouse/ClickHouse/pull/121383), which only ever passes `chars_format::general`. The bump alone cost about 3% on text float parsing.

The fix makes the rule a compile-time parameter of the parser body. `parse_number_string` keeps its signature and behaviour: it tests the format once and routes to a cold, never-inlined javascript instantiation, so callers that use it directly are unaffected. `from_chars_float_advanced` leaves for a cold `from_chars_float_javascript` as its first statement, before anything else is live, so it compiles to a tail call. Placement matters here: any later, and `options` stays live across the parse and the hot frame picks up two extra spill pairs, which costs as much as the original problem.

### Measurements

Intel Xeon 6975P-C (Granite Rapids), clang 22.1.8.

ClickHouse, instructions in `from_chars_caller<double>::call`: 935 on v8.2.10, 1031 on v8.3.0, 944 on this branch.

ClickHouse, parsing a 20M row text column of `Float64`, `perf stat` for the whole run:

| | instructions | cycles |
| --- | --- | --- |
| v8.2.10 | 16.295G | 3.179G |
| v8.3.0 | 16.437G | 3.274G |
| this branch | 16.354G | 3.182G |

`benchmarks/benchmark.cpp`, cycles per float, best of 7:

| | canada.txt | mesh.txt |
| --- | --- | --- |
| v8.2.10 | 47.03 | 24.78 |
| v8.3.0 | 47.34 | 26.08 |
| this branch | 46.28 | 24.86 |

`ctest` is green with clang 22 and with gcc in C++20 mode. `clang-format-17 --style=file` reports no changes.


lemire:
@Algunenano This looks good, running tests, and then I will merge.

PR #414, opened 2026-09-23
Avoid branch mispredictions on the exponent sign and out-of-range results

Algunenano:
The sign of the exponent is as unpredictable as the input, and for out-of-range values it also decides between underflow and overflow. Three places branch on it; this makes them branch-free.

- Exponent sign: parsed without a branch. It uses `|` rather than `||`: GCC turns the `||` into a branch on "has a sign", which is the same bit when positive exponents are written without `+`.
- `compute_float`: both ends of the range are checked with one unsigned comparison, and zero vs infinity is selected without a branch.
- Out-of-range test in `from_chars_float_advanced`: still one branch marked unlikely, as in #413, but tested with one comparison (the exponent is zero or infinite), and the error is selected inside it. Clang otherwise splits the condition into one jump per case.

All three are needed: removing only the first moves the misprediction to the other two for `float`, where most of these values go out of range.

Found with `perf record -b` on exponent-heavy input: the exponent sign test was the most mispredicted branch in the parser.

### Measurements

`benchmarks/benchmark.cpp` on an AMD Ryzen 9 7950X3D (Zen 4), clang 22.1.8 and GCC 16.2.1, cycles and branch misses per float, one run each (the benchmark repeats 1000 times).

HIGGS is real data in NumPy's default `%.18e` format: the first 200k values of the [UCI HIGGS dataset](https://archive.ics.uci.edu/dataset/280/higgs) (first 100k rows, one value per line). Every value has an exponent, with `e-01` and `e+00` about evenly split.

**HIGGS**

| | clang cycles | clang misses | GCC cycles | GCC misses |
| --- | --- | --- | --- | --- |
| ASCII 64-bit | 89.33 → 83.16 (1.07x) | 1.10 → 0.71 | 84.57 → 77.71 (1.09x) | 1.10 → 0.71 |
| ASCII 32-bit | 91.55 → 83.81 (1.09x) | 1.09 → 0.71 | 84.14 → 79.65 (1.06x) | 1.10 → 0.71 |
| UTF-16 64-bit | 89.89 → 84.31 (1.07x) | 1.09 → 0.71 | 86.45 → 78.53 (1.10x) | 1.10 → 0.70 |
| UTF-16 32-bit | 90.06 → 83.98 (1.07x) | 1.09 → 0.71 | 86.17 → 80.79 (1.07x) | 1.09 → 0.70 |

Instructions per float are equal or lower in all eight configurations.

**canada**

| | clang cycles | clang misses | GCC cycles | GCC misses |
| --- | --- | --- | --- | --- |
| ASCII 64-bit | 51.40 → 49.86 (1.03x) | 0.11 → 0.11 | 45.16 → 42.58 (1.06x) | 0.11 → 0.11 |
| ASCII 32-bit | 52.87 → 50.29 (1.05x) | 0.10 → 0.10 | 44.98 → 43.15 (1.04x) | 0.10 → 0.10 |
| UTF-16 64-bit | 58.08 → 56.82 (1.02x) | 0.11 → 0.11 | 52.00 → 49.16 (1.06x) | 0.11 → 0.11 |
| UTF-16 32-bit | 59.13 → 57.00 (1.04x) | 0.10 → 0.10 | 52.71 → 49.58 (1.06x) | 0.10 → 0.10 |

canada has no exponents, so branch misses do not change; the gain comes from the shorter range and out-of-range checks, which run for every value that misses Clinger's fast path (instructions per float, ASCII 64-bit: clang 244.6 → 242.9, GCC 212.4 → 206.5).

**mesh** does not reach the changed code (its values take Clinger's fast path). Its differences, 0.99x to 1.09x, move with code alignment: with `-falign-functions=64 -falign-loops=64` or `-mllvm -align-all-nofallthru-blocks=5` they change in both directions while instructions per float stay the same. They are layout, which is why I have not included them here.


include/fast_float/ascii_number.h
```diff
@@ -491,14 +491,13 @@ parse_number_string_impl(UC const *p, UC const *pend,
         (UC('D') == *p)) {
       ++p;
     }
+    // No branch on the sign: it is as unpredictable as the value (`|`, not
```
lemire: **[suggestion]** The new comment narrates the rewrite and embeds compiler history (`|` rather than `||` because GCC branches). The non-obvious constraint is real, but the comment is longer than it needs to be and records how one compiler lowers `||` today.

**Suggestion:** Keep a short reason, for example that the exponent sign is unpredictable so the advance must be `|` (0 or 1), not `||`. Drop the GCC note.

include/fast_float/decimal_to_binary.h
```diff
@@ -102,15 +102,14 @@ template <typename binary>
...
-    answer.power2 = 0;
-    answer.mantissa = 0;
-    // result should be zero
-    return answer;
-  }
-  if (q > binary::largest_power_of_ten()) {
-    // we want to get infinity:
-    answer.power2 = binary::infinite_power();
+  // One comparison for both ends of the range, then zero vs infinity without
+  // a branch: it follows the sign of the exponent, which is unpredictable.
```
lemire: **[suggestion]** The comment restates the branchless range check and says zero vs infinity "follows the sign of the exponent." That is not what the code does. Inside the early return, `w == 0` forces a zero even when `q > largest_power_of_ten()` (same as the old first branch). Only a nonzero mantissa selects infinity, and only when `q` is above the range rather than below it. A later edit that dropped the `w == 0` term would turn `0e999` into infinity.

**Suggestion:** If a comment stays, say that a zero mantissa is zero for any `q`, and a nonzero mantissa below the table is zero while one above it is infinity. Do not describe that as the sign of the exponent.

include/fast_float/parse_number.h
```diff
@@ -455,14 +455,15 @@ from_chars_float_advanced(UC const *first, UC const *last, T &value,
...
-  // branch instead of folding it into a chain of conditional moves that
-  // every conversion pays for.
-  if fastfloat_clang_unlikely ((pns.mantissa != 0 && am.mantissa == 0 &&
-                                am.power2 == 0) ||
-                               am.power2 ==
-                                   binary_format<T>::infinite_power()) {
-    answer.ec = std::errc::result_out_of_range;
+  // Test for over/underflow. One comparison catches both a zero and an
+  // infinite exponent, so normal values pay a single, never-taken branch. The
+  // result is then selected: which case it is follows the sign of the exponent.
```
lemire: **[suggestion]** The replacement comment restates the comparison and claims two things the predicate does not do. `uint32_t(am.power2 - 1) >= max_power2` is true for every `power2 == 0`, not only a zero result: subnormals (`power2 == 0`, nonzero mantissa) and exact zeros that miss Clinger (for example `0e23`) enter this "never-taken" branch and then correctly leave `answer.ec` alone. The inner selection is also not the sign of the exponent; overflow is `power2 != 0` (on this path, `infinite_power()`), and underflow is a zero mantissa with a nonzero `pns.mantissa`. The code handles those cases; the comment does not.

**Suggestion:** Describe the branch as "exponent field is 0 or infinite" and the inner choice as overflow versus underflow-to-zero, explicitly excluding subnormals and an exact zero. Drop the benchmark/branch narrative.

lemire:
@Algunenano Good in principle, but I am getting register pressure under GCC with this PR. We may need to ease register pressure before this can be safely applied.

Algunenano:
I've verified locally and reproduced the register pressure with gcc 14 and 15, but not 16 (the one I use by default).

I've let Claude analyze it and it appears that, for some reason, gcc decides to load the sign character to a register and keep it alive Gcc decides to store the exact character into a register and keep it alive during the digit loop, so later on it can compute neg_exp from it.

It seems this is already avoided by 1eac92b which is part of https://github.com/fastfloat/fast_float/pull/416

There is a similar (pre-existing) issue with the number's first character where it's kept alieve in a register to compute `answer.negative` later (even in the branches where it's not used), instead of keeping a bool.

lemire:
@Algunenano Could we finish https://github.com/fastfloat/fast_float/pull/416 before we come back this issue here?

Algunenano:
@lemire If you are ok with a longer PR (in terms of commits), I can close this and work fully on https://github.com/fastfloat/fast_float/pull/416

lemire:
@Algunenano Yeah, I think it is worth doing. Closing.

PR #415, opened 2026-09-23
Average branch misses over the repetitions in the benchmark

Algunenano:
`bm/f` summed branch misses over all repetitions without dividing by their count, unlike the other counters, so it printed about 1000 times the misses per float (e.g. 108 instead of 0.11 on canada).


lemire:
Merged.

